PR #57, opened 2024-11-19
[CLI]: add more test about checking data in command, data, history table and optimistic lock

baquan1708:
## Description 

This PR incluse these changes:
- test about checking data in command, data, history table
- test optimistic lock
- update github workflow: build package before running test


koichimurakami:
Thank you!

PR #83, opened 2024-12-02
[Sequence} Update generateSequenceItem function parameters in Sequence package

quangquandev:
### Change the options from a required parameter to an optional parameter.

https://github.com/mbc-net/mbc-cqrs-serverless-doc/pull/10


koichimurakami:
Thank you!

PR #400, opened 2026-04-09
refactor(import): remove ZipImportQueueEventHandler and streamline import processing

baquan1708:
## Description

This PR transitions the ZIP import pipeline from an asynchronous, queue-based architecture (SNS -> SQS -> Lambda) into a highly efficient, synchronous API flow. By directly orchestrating the Step Function from the ImportService, we provide real-time feedback to the API client, eliminate unnecessary infrastructure hops, and capture rich HTTP audit context.


koichimurakami (commented):
## Review Summary

Thank you for the refactor. The direction of moving ZIP processing into a synchronous API flow is well-reasoned, and the test improvements in `import.service.spec.ts` are a clear upgrade. However, there are several issues that need to be addressed before merging.

---

### [Critical] S3 key format (13 digits) does not match the regex in `ZipImportSfnEventHandler` (14 digits)

This is the most severe bug in this PR. The two pieces of code are now incompatible with each other.

**`import.service.ts:206` — key generated with `Date.now()` (13 digits):**
```typescript
const extractedKey = `csv-imports/${dto.tenantCode}/${Date.now()}-${filename}`
// e.g. csv-imports/tenant1/1775740817220-contract.csv
//                           ^^^^^^^^^^^^^ 13 digits (Unix milliseconds)
```

**`zip-import.sfn.event.handler.ts:65` — regex expects 14 digits (`yyyyMMddHHmmss`):**
```typescript
const match = s3Key.match(/\d{14}-(.+)\.csv$/)
// e.g. "1775740817220-contract.csv" → NO MATCH
```

`Date.now()` always returns a 13-digit timestamp. The regex was written for a `yyyyMMddHHmmss` format (14 digits). They never match.

**Impact:** When `dto.tableName` is omitted (the DTO's default is `null`), the `trigger_single_csv_and_wait` SFN state always throws:
```
Error: Could not parse tableName from filename: csv-imports/tenant1/1775740817220-contract.csv.
       Expected format: yyyymmddhhMMss-{tableName}.csv
```

The old `ZipImportQueueEventHandler` passed the **original filename** from inside the ZIP to the SFN, so user-defined naming conventions like `yyyyMMddHHmmss-tableName.csv` worked. The new implementation prepends `Date.now()-` to the filename, breaking that convention.

**Suggested fix (Option A):** Use the job's `taskCode` (ULID) as the S3 prefix instead of `Date.now()`, so the original filename is preserved as the last path segment:
```typescript
const extractedKey = `csv-imports/${dto.tenantCode}/${taskCode}/${filename}`
// → s3Key passed to SFN: "csv-imports/tenant1/<ulid>/20250918120000-contract.csv"
// → regex \d{14} matches the original filename correctly
```

**Suggested fix (Option B):** Fix the regex to match 13-digit timestamps:
```typescript
const match = s3Key.match(/\d{13}-(.+)\.csv$/)
```

---

### [High] `ConfigService` is injected twice into the same constructor

```typescript
// import.service.ts:49, 54
constructor(
  ...
  private readonly config: ConfigService,        // existing
  ...
  private readonly configService: ConfigService, // newly added (duplicate)
  private readonly sfnService: StepFunctionService,
) {
  this.alarmTopicArn = this.config.get<string>('SNS_ALARM_TOPIC_ARN')    // uses "config"
  ...
  const sfnArn = this.configService.get<string>('SFN_IMPORT_ZIP_ORCHESTRATOR_ARN') // uses "configService"
```

NestJS injects the same instance into both parameters, so this works at runtime — but it is a clear code smell and will confuse future maintainers. Please remove `configService` and use the existing `config` field throughout.

---

### [High] `@HttpCode(202)` contradicts synchronous blocking behavior

The controller declares HTTP 202 Accepted, which semantically means "the request has been accepted for **asynchronous** processing." But when `sortedFileKeys` is not provided, `createZipJob()` runs the full synchronous pipeline in-process: S3 download → in-memory ZIP extraction → per-file S3 upload → SFN start execution.

```typescript
// import.controller.ts:104
@HttpCode(HttpStatus.ACCEPTED) // implies async / fire-and-forget
async createZipImport(...) {
  return this.importService.createZipJob(...) // actually blocks until SFN is started
}
```

For large ZIP files this pipeline can exceed the API Gateway timeout (29 s) or the Lambda memory limit. Please clarify the design intent:
- If the operation is truly synchronous (small files only), return `200 OK` and document the size limits.
- If large files must be supported, keep the original async queue approach for the extraction step.

---

### [Medium] Empty `sortedFileKeys: []` silently changes behavior

The old `ZipImportQueueEventHandler` explicitly rejected an empty array:
```typescript
// old behavior
if (sortedFileKeys && sortedFileKeys.length === 0) {
  throw new Error('Sorted file keys are required.')
}
```

The new implementation silently falls through to the S3 download path:
```typescript
// new behavior
let sortedS3Keys = dto.sortedFileKeys || []
if (sortedS3Keys.length === 0) {
  // → triggers S3 download + extraction
}
```

A caller that passes `sortedFileKeys: []` expecting an error will instead trigger an unintended S3 download. Consider restoring the explicit guard or at least documenting this behavior change.

---

### [Medium] `updateImportJob` overwrites the entire `attributes` field

```typescript
// import.service.ts:225-227
await this.updateImportJob(masterJobKey, {
  set: { attributes: { extractedFileKeys: extractedKeys } },
  // ← dto fields (bucket, key, tableName, tenantCode) are lost
})
```

`DynamoDbService.updateItem` with `SET #attributes = :attributes` replaces the entire `attributes` map. The old handler spread the original DTO:
```typescript
// old: attributes preserved
set: { attributes: { ...zipJobAttributes, extractedFileKeys } }
```

The SFN flow reads from `event.context.Execution.Input`, not from DynamoDB, so this does not break processing. But the DynamoDB record loses its original context, making debugging and auditing harder.

---

### [Low] `ZipImportSfnEventHandler` tests deleted with no replacement

Both `zip-import.queue.event.handler.spec.ts` and `zip-import.sfn.event.handler.spec.ts` have been removed. `ZipImportSfnEventHandler` still exists and contains non-trivial logic (`trigger_single_csv_and_wait`, `finalize_zip_job`, hook execution). This is also why the Critical regex bug above was not caught — there are no tests that exercise the `tableName`-from-filename path with the new S3 key format.

Please add tests for at least:
- `trigger_single_csv_and_wait` with `tableName` provided
- `trigger_single_csv_and_wait` with `tableName` omitted (exercises the regex path)
- `finalize_zip_job` with mixed COMPLETED/FAILED results

PR #433, opened 2026-05-12
feat: add AppSync Events API support for CQRS notification flow

baquan1708:
## Summary
 
Adds support for **AWS AppSync Events API** as a pluggable pub/sub transport for the CQRS notification flow. Transport selection is controlled via the `NOTIFICATION_TRANSPORTS` env var.

---

## Channel Design

 
Channel structure mirrors the GraphQL subscription filter variables so the client switches transport without changing filter logic:
 

```
/{namespace}/{tenantCode}/{action}/{sanitizedId}
    seg 1       seg 2      seg 3       seg 4

```

 
Non-alphanumeric characters in `id` (e.g. `#`, `@` from DynamoDB pk/sk) are sanitized to `-`.
 

| Client goal | Subscribe to |
| --- | --- |
| All events for tenant | `/{namespace}/{tenantCode}/*` |
| Filtered by action | `/{namespace}/{tenantCode}/command-status/*` |
| Track one specific command | `/{namespace}/{tenantCode}/command-status/{sanitizedId}` |
|   |  |

---

## Transport Selection & Custom Registration

 
Determine which discovered transports actually receive broadcasted messages using the environment variable:

```bash
NOTIFICATION_TRANSPORTS=appsync-graphql                # default, backward compatible
NOTIFICATION_TRANSPORTS=appsync-event                  # Events API only
NOTIFICATION_TRANSPORTS=appsync-graphql,appsync-event  # dual-publish / migration

```


koichimurakami:
## Code Review

### 🔴 Critical

**1. `apiKeys?.[0]` is always `undefined` — API key auth silently falls back to IAM**

`EventApi.apiKeys` is `{ [key: string]: CfnApiKey }` (an object, not an array), so `[0]` always resolves to `undefined`. The `?? ''` fallback then injects an empty string into `APPSYNC_EVENTS_API_KEY`. Since `''` is falsy, `AppSyncEventsService.postToChannel` skips the API key branch and falls through to IAM SigV4 — silently, with no error. As long as IAM permissions are correct the feature still works, but the intended auth mode is never used.

```typescript
// infra-stack.ts L323, L488 — current (broken)
appSyncEventsApi.apiKeys?.[0]?.attrApiKey ?? ''

// fix — key name is 'Default' when apiKeyConfig.name is not set
appSyncEventsApi.apiKeys['Default']?.attrApiKey ?? ''
// or
Object.values(appSyncEventsApi.apiKeys)[0]?.attrApiKey ?? ''
```

---

**2. `namespace` being `undefined` causes `cdk synth` to fail**

`Config.appsyncEvents.namespace` is `string | undefined`, but `CfnOutput.value` and the Lambda environment map both require `string`. With `"strictNullChecks": true` in `infra/tsconfig.json`, this is a compile error that blocks `cdk synth`.

```typescript
// infra-stack.ts L327, L489 — fix
value: props.config.appsyncEvents.namespace ?? 'default',
APPSYNC_EVENTS_NAMESPACE: props.config.appsyncEvents.namespace ?? 'default',
```

---

**3. `Promise.allSettled` swallows errors all the way up the call stack**

`Promise.allSettled` never rejects, so when `NotificationEventHandler.execute` returns it, `EventBus.execute` (which wraps handlers in `Promise.all`) sees a resolved value and the error is completely lost — not logged, not propagated to `EventService`.

```typescript
// notification.event.handler.ts — fix
const results = await Promise.allSettled(tasks)
for (const r of results) {
  if (r.status === 'rejected') {
    this.logger.error('Notification publish failed:', r.reason)
  }
}
return results
```

---

### 🟡 Recommended

**4. Use CDK's `grantPublish()` instead of a manual `PolicyStatement`**

The manual resource ARN `${appSyncEventsApi.apiArn}/*` is broader than necessary. CDK's built-in method uses `apis/{apiId}/channelNamespace/*`, which follows the principle of least privilege and is easier to maintain.

```typescript
// infra-stack.ts L1203-L1210, L1343-L1349 — fix
appSyncEventsApi.grantPublish(lambdaApi)
appSyncEventsApi.grantPublish(taskRole)
```

---

**5. `namespace` is not sanitized in `resolveChannel`**

`tenantCode`, `table`, `action`, and `id` are all passed through `sanitizeSegment`, but `this.namespace` is used as-is. If the env var contains characters invalid for an AppSync channel segment, the publish call returns a 400 error at runtime.

```typescript
// appsync-events.service.ts
return `/${this.sanitizeSegment(this.namespace)}/${tenantCode}/...`
```

---

**6. `notification.event.handler.spec.ts` is missing from the diff**

The PR description mentions a new spec for `NotificationEventHandler` covering all four dispatch modes (events disabled / events-only / dual-publish / string `'true'` vs boolean `true`), but the file is not in the actual diff. Please add it — these branches are the core of the migration logic.

---

### 🟢 Minor

**7. `appSyncEventsNamespace` is an unused variable**

The `ChannelNamespace` construct is registered with the CDK scope via `new`, so the variable reference itself is not needed. `noUnusedLocals` is `false` in `tsconfig.json` so this won't break CI, but it adds noise.

```typescript
// infra-stack.ts — simplify
new cdk.aws_appsync.ChannelNamespace(this, 'events-namespace', { ... })
```

koichimurakami (commented):
## Code Review

Great PR overall — the pluggable transport architecture is clean and the test coverage is solid. A few items worth addressing before merge.

---

### Must Fix

**README.md rendering is broken** (`examples/appsync-event-client/README.md`)

The file starts with an AI-generated meta sentence and the entire content is wrapped in a ````markdown```` fenced block, so GitHub renders it as a raw code block instead of formatted markdown. Please remove the meta text and unwrap the content.

Also: the file is missing a newline at EOF (`.gitignore` has the same issue).

---

### Recommended

**Missing guard in `AppSyncService.sendMessage`** (`packages/core/src/notifications/appsync.service.ts`)

When `APPSYNC_ENDPOINT` is not set, `this.hostname` and `this.signer` remain `undefined`. The constructor now wraps initialization in `if (this.endpoint)`, but `sendMessage` has no corresponding guard, so calling it without a configured endpoint throws a silent `TypeError`. `validateBuiltinNotificationTransportEnv` protects production, but direct instantiation in tests or custom setups will hit this. A short early-return or explicit check at the top of `sendMessage` would make this self-defending.

**CDK API key retrieval is order-dependent** (`packages/cli/templates/infra/libs/infra-stack.ts`)

```typescript
Object.values(appSyncEventsApi.apiKeys)[0] as cdk.aws_appsync.CfnApiKey
```

`Object.values()` order is implementation-defined. If `EventApi` ever generates more than one key internally, the wrong key could end up in the output. Retrieving the key by its construct ID or logical name would be more deterministic.

---

### Minor

**Hardcoded placeholder in `page.tsx`** (`examples/appsync-event-client/src/app/page.tsx`)

```typescript
apiKey: 'da2-xxxxxxxxxxxxxxxxxxxxxxxxxx', // <-- UPDATE WITH YOUR REAL API KEY
```

The README instructs developers to edit `page.tsx` directly, which makes it easy to accidentally commit a real key. Reading from `process.env.NEXT_PUBLIC_APPSYNC_API_KEY` (with a `.env.local.example` file) would be safer and more idiomatic for Next.js.

koichimurakami:
## Follow-up Review

Good progress on addressing the previous feedback. Here is the updated status.

---

### ✅ Resolved

- **README.md rendering** — meta text removed, markdown unwrapped, properly formatted
- **`.gitignore` EOF newline** — fixed
- **`page.tsx` hardcoded API key** — correctly moved to `NEXT_PUBLIC_*` env vars with a missing-config banner and disabled connect button
- **`AppSyncService.sendMessage` guard** — early-return added when `APPSYNC_ENDPOINT` is not set
- **`AppSyncEventsService.sendMessage` guard** — early-return added (`!this.url || !this.signer`)
- **CDK `grantPublish()`** — now uses the CDK method instead of a manual policy statement

---

### 🟢 Minor

**Client-side `sanitize()` operation order differs from the server** (`examples/appsync-event-client/src/lib/subscribe-events.ts`)

The server-side `sanitizeSegment()` truncates to 50 characters first, then strips leading/trailing dashes. The client-side `sanitize()` does it in reverse order. In practice this only produces a mismatch when the `id` field exceeds 50 characters after substitution and the character at exactly position 50 maps to a dash — an edge case unlikely to occur with typical DynamoDB pk/sk values. Still, aligning the client to match the server exactly would make the behavior consistent:

```typescript
private sanitize(value: string): string {
  if (!value) return 'none'
  let s = value.replace(/[^a-zA-Z0-9]/g, '-')
  s = s.substring(0, 50)
  s = s.replace(/^-+|-+$/g, '')
  return s || 'none'
}
```

PR #438, opened 2026-05-22
feat(core): JWT group permissions via custom:groups

baquan1708:
## Summary
- Adds **group-derived roles** to core authorization using a new optional JWT claim **`custom:groups`** (same `{ tenant, role }` JSON shape as `custom:roles`).
- **`RolesGuard`** and cross-tenant header checks use **`tenantRoles`** — the union of direct and group roles for the active tenant (deduplicated).
- **`tenantRole`** is unchanged: still the **first** matching entry from **`custom:roles` only**

## Token contract

| Claim | Meaning |
|-------|---------|
| `custom:roles` | Roles assigned **directly** to the user |
| `custom:groups` | **Flattened** roles from group membership |

- Missing or empty `custom:groups` → same behavior as before
- **`system_admin`** in either claim → full `@Roles` bypass.
- Cross-tenant header override succeeds if **any** `tenantRoles` entry is in `CROSS_TENANT_ROLES`.

## Backward compatibility

- Tokens **without** `custom:groups`: no change for guards.
- **`UserContext.tenantRole`**: same semantics (direct roles only).
- Routes protected by **`@Roles()`** automatically honor group roles.

### DIFF
--- a/packages/core/src/context/index.ts
+++ b/packages/core/src/context/index.ts
@@ -1,2 +1,3 @@
 export * from './invoke'
+export * from './role-resolver'
 export * from './user'
--- a/packages/core/src/context/invoke.ts
+++ b/packages/core/src/context/invoke.ts
@@ -21,6 +21,8 @@ export interface JwtClaims {
   name: string
   'custom:tenant'?: string // tenant code
   'custom:roles'?: string
+  /** JSON array of { tenant, role } — roles granted via group membership (flattened at token issuance). */
+  'custom:groups'?: string
   exp: number
   email: string
   email_verified?: boolean
--- a/packages/core/src/context/role-resolver.spec.ts
+++ b/packages/core/src/context/role-resolver.spec.ts
@@ -1,0 +1,82 @@
+import { JwtClaims } from './invoke'
+import { parseCustomRolesJson, resolveTenantRoles } from './role-resolver'
+
+describe('parseCustomRolesJson', () => {
+  it('returns empty array when undefined', () => {
+    expect(parseCustomRolesJson(undefined)).toEqual([])
+  })
+
+  it('parses valid JSON array', () => {
+    const json = JSON.stringify([{ tenant: 'T1', role: 'admin' }])
+    expect(parseCustomRolesJson(json)).toEqual([
+      { tenant: 't1', role: 'admin' },
+    ])
+  })
+
+  it('throws on invalid JSON', () => {
+    expect(() => parseCustomRolesJson('not-json')).toThrow()
+  })
+})
+
+describe('resolveTenantRoles', () => {
+  const baseClaims = {
+    sub: 'user-1',
+  } as JwtClaims
+
+  it('unions direct and group roles for tenant', () => {
+    const claims: JwtClaims = {
+      ...baseClaims,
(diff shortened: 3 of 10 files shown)


koichimurakami (commented):
## Code Review

Overall, this is a well-designed PR. The separation between `tenantRole` (direct roles only) and `tenantRoles` (union for authorization) is clean, and backward compatibility is maintained. Test coverage is solid. I have two items that should be addressed before merging, and two minor observations.

---

### 🔴 Must Fix

#### 1. `parseCustomRolesJson` throws on valid but non-array JSON

`JSON.parse` succeeds for inputs like `'null'`, `'"string"'`, or `'{"tenant":"a","role":"b"}'`, but calling `.map()` on the result throws a `TypeError` at runtime:

```
JSON.parse('null')               → TypeError: Cannot read properties of null (reading 'map')
JSON.parse('"string"')           → TypeError: parsed.map is not a function
JSON.parse('{"tenant":"a",...}') → TypeError: parsed.map is not a function
```

If a Cognito admin accidentally sets `custom:groups` to a single object instead of an array, every request for that user will 500. Since `parseCustomRolesJson` is now shared by both `custom:roles` and `custom:groups`, the blast radius is larger than before.

Suggested fix:

```typescript
export function parseCustomRolesJson(json: string | undefined): CustomRole[] {
  if (!json) return []
  try {
    const parsed = JSON.parse(json)
    if (!Array.isArray(parsed)) return []
    return parsed.map((entry) => ({
      ...entry,
      tenant: (entry.tenant || '').toLowerCase(),
    }))
  } catch {
    return []
  }
}
```

The existing test `'throws on invalid JSON'` should be updated to expect `[]` instead of a thrown error, and tests for `'null'` and single-object inputs should be added.

---

#### 2. `getUserRole` override is now silently ignored

Before this PR, `verifyRole` called `this.getUserRole(context)`, so subclasses could override that method to inject custom role resolution logic (e.g., database lookups, dynamic roles). After this PR, `verifyRole` calls `this.getUserTenantRoles(context)` directly, so any subclass that overrides only `getUserRole` will have its logic silently bypassed — no error, no warning.

The `// eslint-disable-next-line @typescript-eslint/no-unused-vars` comment above the method confirms the author already knows it is dead code.

The method should be marked `@deprecated` with a migration note pointing to `getUserTenantRoles`:

```typescript
/**
 * @deprecated Override `getUserTenantRoles` instead. This method is no longer
 * called by `verifyRole` and will be removed in a future release.
 */
protected async getUserRole(context: ExecutionContext): Promise<string> {
  const userContext = getUserContext(context)
  return userContext.tenantRole
}
```

This should also be noted in the PR description / CHANGELOG so consumers know to migrate.

---

### 🟡 Minor observations

#### 3. `getUserContext` is called twice per request

`verifyTenant` calls `getUserContext` directly (line 46), and `verifyRole` calls it again via `getUserTenantRoles` (line 140). The JWT decode runs twice on every request. Not a correctness issue, but worth caching in the execution context if performance becomes a concern.

#### 4. `parseCustomRolesJson` and `resolveTenantRoles` are exported as public API

`context/index.ts` re-exports everything from `role-resolver.ts`, which means these functions become part of the package's public API surface. If they are intended as internal helpers, the export should be removed or the file should be moved to an `internal/` directory to avoid locking the signatures.

Issue #434, opened 2026-05-14
support user group to handle authz


dominhhai:
currently, we only support `custom:roles` to handle the permission for individual.
We want to expands this to a group of users. So that the user in the same group can receive the permission from that group.

PR #440, opened 2026-06-01
feat(core): implement group-based roles and enhance authorization handling

baquan1708:
## Summary
Extends `@mbc-cqrs-serverless/core` authorization so users can inherit roles from **group membership** in the JWT, while keeping existing **direct** roles from `custom:roles`.
- Parses tenant-scoped `custom:groups` in `getUserContext()` → `tenantGroupIds`
- Collects direct roles in `tenantRoles` (group-derived roles are not merged into JWT context)
- `RolesGuard` evaluates: direct `tenantRoles` → optional `@GroupRoleResolver()` → deny
- Apps without a resolver behave exactly as today (direct roles only)
- `@Global() AuthModule` registers `GroupRoleResolverRegistry` so `RolesGuard` in feature modules receives the same singleton
## Motivation
Today only `custom:roles` drives `@Auth()` / `@Roles()`. Product teams need group membership in the token (`custom:groups`) and app-owned storage for group → role mappings (not in the JWT).

### JWT / context
| Claim / field | Behavior |
|---------------|----------|
| `custom:groups` | New optional claim; tenant-scoped `groups[]`; missing → `tenantGroupIds = []` |
| `custom:roles` | Unchanged shape; drives `tenantRoles` + existing `tenantRole` selection |
| `UserContext` | Adds `tenantRoles`, `tenantGroupIds` |
### Authorization
| Component | Responsibility |
|-----------|----------------|
| `@GroupRoleResolver()` | Marks exactly one app resolver class; discovered at bootstrap |
| `GroupRoleResolverRegistry` | Holds resolved DI instance |
| `AuthzBootstrapService` | `OnApplicationBootstrap`: 0 or 1 resolver; **throws** if >1 or if type discovered but DI cannot resolve instance |
| `RolesGuard` | `system_admin` bypass from **direct** `tenantRoles` only; group path fail-closed on resolver errors |
## Evaluation order
1. No `@Roles` metadata → allow  
2. `system_admin` in `tenantRoles` → allow  
3. Any required role in `tenantRoles` → allow  
4. If registry + `tenantGroupIds` → `resolveRoles({ tenantCode, groupIds, claims })`  
5. Any required role in returned group roles → allow; else deny  
## Related Issues
- **Closes** #[434](https://github.com/mbc-net/mbc-cqrs-serverless/issues/434)


dominhhai:
@koichimurakami 
ユーザーグループでのロール制御を追加しました。
ご確認のほどよろしくお願いいたします。
なお、ファイル管理などのグループはまだ設計中ですがファイルパッケージで対応する予定です。

koichimurakami:
## Code Review: Group-based roles & authorization

Thanks for this PR! The overall direction (group-derived roles via `IGroupRoleResolver`) looks solid. I went through the changes carefully and have a few findings, ordered by severity. Happy to discuss any of them.

### 🔴 High

**1. Malformed `custom:groups` can crash every authenticated request path**
`packages/core/src/context/user.ts` — `parseTenantGroups` does `JSON.parse(raw || '[]').map(...)` with no error handling. Two inputs throw:
- Invalid JSON (`'invalid'`) → `SyntaxError`
- Valid-but-not-an-array JSON (`'{"a":1}'`, `'true'`, `'5'`) → `TypeError` on `.map`

`getUserContext` is called well beyond `RolesGuard` — notably `request-logger.service.ts` (runs on **every** Lambda log line, regardless of route guards), and the command write/read paths (`command.service.ts`, `repository.ts`), none of which wrap it in try/catch. Since `custom:groups` is a brand-new, app-populated claim, a single malformed token would turn previously-working requests into 500s — even on endpoints with no `@Roles`.
**Suggestion:** fail closed in `parseTenantGroups` (try/catch + `Array.isArray` → `[]`), and guard the logger so enriching a log line can never break logging itself.

### 🟠 Medium

**2. `getUserRole()` overrides are now silently ignored**
`packages/core/src/guard/roles.guard.ts` — `verifyRole` no longer calls `getUserRole()`, but the `protected` method remains. Subclasses that override it to inject custom authorization will compile and run, yet never be invoked — a silent behavior change. Consider removing it (to surface the break) or routing `verifyRole` through it again.

**3. Role-name matching is case-sensitive while tenant matching is not**
`hasAnyRole` / the `ROLE_SYSTEM_ADMIN` check compare exact strings. Tenants are lowercased but role names are not, so `'Admin'` vs `@Roles('admin')`, or `'SYSTEM_ADMIN'` vs `'system_admin'`, would mismatch. Suggest normalizing role names to lowercase in `getUserContext` for consistency.

**4. Resolver failures are swallowed and surface as 403**
`resolveGroupRoles` catches resolver errors and returns `[]`. A transient backend outage then looks identical to a genuine "access denied", and group-only users get mass 403s with no signal. Consider rethrowing so the request returns a 5xx, distinguishing availability failures from authorization decisions.

**5. Resolver DI scope is not validated**
`authz-bootstrap.service.ts` fetches the resolver once at bootstrap via `moduleRef.get(..., { strict: false })`. A `REQUEST`/`TRANSIENT`-scoped resolver would be frozen to a context-less instance and silently misbehave per request. Suggest rejecting non-singleton scopes at bootstrap and documenting that the resolver must be stateless/singleton.

### 🟡 Low (cleanup)

**6.** `resolveGroupRoles` re-derives `userContext` via `getUserContext(context)` although `verifyRole` already computed it — this re-parses the JWT twice on the hot path. Pass the computed `userContext` in.
**7.** `index.ts` re-exports `AuthzBootstrapService` (an internal `OnApplicationBootstrap` hook not even in `AuthModule`'s `exports`), leaking it into the public API. Consider excluding it.
**8.** `@GroupRoleResolver()` already applies `@Injectable()`, but the test code shows `@GroupRoleResolver() @Injectable()` together. Worth a JSDoc note ("don't add `@Injectable()` yourself") to avoid accidental scope overrides.

### ✅ Looked at and OK
- `canOverrideTenant` (single `tenantRole`) vs `verifyRole` (`tenantRoles` array) cannot diverge — `tenantRole ∈ tenantRoles` always holds and `verifyTenant` gates `verifyRole`.
- `undefined` tenantCode never reaches the resolver — `verifyTenant` early-returns and the empty-`groupIds` check both block it.
- Re: additive global roles in `collectTenantRoles` — confirmed offline that global roles are `system_admin`-only in practice, so there's no real privilege change vs. the old single-role logic. Just noting the "global = admin-only" assumption isn't enforced in code, in case that changes later.

koichimurakami:
## Re-review after `c8cf23e` — thanks for the quick turnaround!

Re-reviewed against the latest commit (`feat(core): enhance group role resolver and improve error handling`). The high/medium findings are handled well, all 98 unit tests pass, and a couple of items need a correction from me. Details below.

### ✅ Resolved
- **Malformed `custom:groups`** — `parseTenantGroups` now fails closed (try/catch + `Array.isArray`, plus an `Array.isArray(groups)` guard on the result). Tests updated to assert empty `tenantGroupIds` instead of a throw, including a non-array case. 👍
- **Resolver failures** — now rethrows instead of returning `[]`, so a backend outage surfaces as 5xx rather than a silent 403. Test renamed accordingly. 👍
- **`getUserRole()`** — now `@deprecated` with a clear pointer to override `verifyRole`/`resolveGroupRoles`/`canOverrideTenant`, plus an "Extending authorization" note in the README. 👍
- **`@GroupRoleResolver()` double-`@Injectable`** — JSDoc + README now warn against it. 👍

### ✏️ Correction on my earlier point — please disregard the "case-sensitive roles" finding
I flagged that role names aren't lowercased. After re-reading `backward-compatibility.spec.ts` ("should handle mixed case roles from legacy systems" asserts `tenantRole === 'Admin'`), I see that **preserving role-name case is intentional** for legacy compatibility — only `tenant` is normalized, by design. Lowercasing roles would break that contract. **My bad — no change needed here.** (At most, a docs note that apps must match role casing between `custom:roles` and `@Roles()`.)

### 🟡 Optional follow-ups (not blockers)
These are nice-to-haves, not merge blockers — listing for completeness:
1. **Same fail-closed treatment for `custom:roles`** (`user.ts`) — `custom:groups` is now guarded, but the `custom:roles` `JSON.parse` right above it still isn't. This is pre-existing behavior (not introduced here), but extending the same guard would make the function consistent.
2. **`request-logger.service.ts`** still calls `getUserContext` unguarded. With `custom:groups` fixed the main crash vector is gone; wrapping it would just ensure log enrichment can never break logging itself (e.g. on a malformed token).
3. **`resolveGroupRoles` re-derives `userContext`** (`roles.guard.ts`) — minor double JWT parse on the group path; could pass the already-computed `userContext` in. Negligible cost.
4. **`index.ts` re-exports `AuthzBootstrapService`** — internal bootstrap hook; could be kept out of the public surface. No real misuse today.
5. **Resolver scope** is documented but not enforced; an explicit bootstrap check rejecting non-singleton scopes would be belt-and-suspenders. Docs are probably enough.

Overall this looks good to me — the substantive issues are addressed. The remaining items are polish and can be deferred. 🚀

PR #443, opened 2026-06-03
fix(directory): change table name

tudang-z:


koichimurakami (changes requested):
## Review: incomplete rename leaves a confirmed history-table bug

This PR renames the directory package's DynamoDB PK prefix (`DIRECTORY` → `DOCUMENT`) and the `CommandModule` table name (`directory` → `document`). The rename is incomplete in one place, which causes a confirmed runtime bug.

### 🔴 Confirmed bug — history table name not renamed

`packages/directory/src/directory.service.ts:455`

```ts
const table = `${process.env.NODE_ENV}-${process.env.APP_NAME}-directory-history`
```

After this PR, `CommandModule.register({ tableName: 'document' })` makes the core `HistoryService` write version history to `<NODE_ENV>-<APP_NAME>-document-history` (`HistoryService` derives the name via `getTableName(options.tableName, TableType.HISTORY)` → `<prefix><moduleName>-history`). But `findHistory()` still reads from the hardcoded `directory-history` table.

**Failure scenario:** new version-history records are written to `document-history`, while `findHistory()` queries `directory-history` (the old / nonexistent table) via `listItemsByPk`. As a result, history listings always come back empty or fail with `ResourceNotFoundException`. The unit test does not catch this because `DynamoService.listItemsByPk` is mocked.

`directory-history` is the only place this old name remains (single occurrence in the diff). **Fix:** `directory-history` → `document-history`.

### 🟡 Naming inconsistency — `prismaService.directory`

`packages/directory/src/directory.service.ts:120`

```ts
const fileSizeSummary = await this.prismaService.directory.groupBy({ ... })
```

This is **not** a functional bug — the Prisma model name is supplied by the consumer app and is not coupled to the DynamoDB table name, so the framework still works as long as the consumer keeps a `directory` model. However, it breaks the convention used elsewhere (e.g. the `master` package uses `tableName: 'master'` + `prismaService.master`), where the table name and the Prisma model name match. Since the DynamoDB table is now `document`, leaving `prismaService.directory` is inconsistent. Consider aligning the names, or documenting that the consumer's Prisma model must stay named `directory` regardless of the table name.

### ⚠️ No migration strategy

Renaming both the PK prefix and the table name is a **breaking change**: existing deployed data (`DIRECTORY#tenant` PKs in the `directory-*` tables) becomes unreachable after deploy. The commit is labeled `fix`, but this is breaking. Please confirm whether a data migration is required, or whether this feature has no production data yet. Note that `restoreHistoryItem()` reuses the stored `item.id`, so restoring pre-existing data would mix old (`DIRECTORY#...`) and new IDs.

### ✅ Looks good

- The controller and `directory-file.service` don't build PKs independently, so they stay consistent with the service switching to `DOCUMENT#`.
- The spec fixture updates are consistent with the source changes.

**Must-fix before merge:** the `directory-history` → `document-history` change at line 455.

PR #466, opened 2026-07-07
fix(infra): lower ATTRIBUTE_LIMIT_SIZE default for Step Functions payload limit

baquan1708:
## Summary

- Lower the framework default `ATTRIBUTE_LIMIT_SIZE` from `389120` (380 KB) to `102400` (100 KB) in CDK infra templates and env examples
- The previous default was sized for DynamoDB’s 400 KB item limit, but Step Functions enforces a 256 KB payload limit per state
- Command state machines pass the DynamoDB stream event twice (`input.$` + `context.$` / `$$.Execution.Input`), so inline attributes above ~110 KB can trigger `States.DataLimitExceeded` (observed at `check_version` in production)
- Add inline comments at each definition site explaining the Step Functions constraint and why DynamoDB’s limit should not be used as the reference
- Update CDK snapshot to reflect the new Lambda env default
## Problem

Production error:

```
error: States.DataLimitExceeded
cause: The state/task 'check_version' provided parameters with a size
       exceeding the maximum number of bytes service limit.
```

**Root cause:** `ATTRIBUTE_LIMIT_SIZE` allowed attributes up to 380 KB inline in DynamoDB, but Step Functions payloads are approximately:

```
payload ≈ DynamoDB event × 2 + ~20 KB context overhead
```

Safe inline limit: `(256 KB − overhead) ÷ 2 ≈ 110 KB`. The old default was ~3.2× too high.

## Solution

When `attributes` exceed `ATTRIBUTE_LIMIT_SIZE`, the framework already offloads them to S3 via `DynamoDBService.objToDdbItem()`. This PR lowers the default so offload happens earlier, keeping Step Functions payloads under 256 KB.

**Default value rationale:**

```
SFN limit:           262,144 bytes (256 KB)
Context overhead:    ~20,480 bytes (~20 KB)
Duplication factor:  ×2
Safety margin:       ~8,192 bytes (~8 KB)
Rounded default:     102,400 bytes (100 KB)
```

### DIFF
--- a/packages/cli/templates/.env.local
+++ b/packages/cli/templates/.env.local
@@ -48,8 +48,12 @@ LOCAL_DDB_ADMIN_PORT=8001
 # DynamoDB endpoint, useful for local development
 DYNAMODB_ENDPOINT=http://localhost:${LOCAL_DYNAMODB_PORT:-8000}
 DYNAMODB_REGION=ap-northeast-1
-# set the limit size for `attributes` of object in DDB
-ATTRIBUTE_LIMIT_SIZE=389120 # bytes, refer to https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ServiceQuotas.html#limits-attributes
+# Max size (bytes) for `attributes` before S3 offload.
+# Must account for Step Functions 256 KB payload limit: the SFN state passes
+# the DynamoDB event twice (input + context.Execution.Input), so the safe
+# limit is ~(256 KB - overhead) / 2 ≈ 110 KB. Default: 100 KB.
+# DynamoDB item limit is 400 KB — do NOT use that as the reference.
+ATTRIBUTE_LIMIT_SIZE=102400
 # S3 endpoint, useful for local development
 S3_ENDPOINT=http://localhost:${LOCAL_S3_PORT:-4566}
 S3_REGION=ap-northeast-1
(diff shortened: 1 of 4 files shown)


koichimurakami:
This PR lowers `ATTRIBUTE_LIMIT_SIZE` to 100KB to avoid the Step Functions 256KB payload limit, but I think the underlying Step Functions payload design should be reconsidered.

Currently, `DataSyncNewCommandEventHandler` passes the whole DynamoDB Stream record directly as the Step Functions execution input. Downstream states then read `dynamodb.NewImage` from `context.Execution.Input` and call `unmarshall()`. This means Step Functions carries the entire DynamoDB Streams event, including the AttributeValue representation, ARN, Keys, SequenceNumber, and other metadata, rather than just the business data. On top of that, each state passes both `input.$: "$"` and `context.$: "$$"`, so the original execution input is duplicated inside the LambdaInvoke payload.

With this structure, having raw `attributes` under 100KB is not necessarily safe. The actual Step Functions payload contains the whole DynamoDB Stream record, not just `attributes`, and `NewImage` is expanded into DynamoDB AttributeValue objects. For attributes with many small keys, that representation can be much larger than the raw JSON. So the 100KB threshold can still exceed the 256KB Step Functions limit depending on the shape of the attributes.

The downstream logic seems to need mainly `commandTableName` and `pk/sk`, or the unmarshalled CommandModel. If so, would it be better to normalize the Step Functions input at the start instead of passing the raw DynamoDB Stream record? For example, we could pass either the unmarshalled business JSON from `NewImage`, or an even smaller payload such as `commandTableName` plus `commandKey`.

That would avoid the payload expansion caused by DynamoDB AttributeValue objects and unnecessary Stream metadata, and should make the design more robust against the Step Functions payload limit than only lowering the default `ATTRIBUTE_LIMIT_SIZE`.

PR #503, opened 2026-08-20
feat(infra): swap LocalStack for Floci as the local S3 emulator

baquan1708:
## Summary

LocalStack Community Edition reached end of life in March 2026: it now requires an auth token and no longer receives security updates. Our local stack used LocalStack for **exactly one** service — S3 — while every other AWS dependency already has a dedicated emulator (`dynamodb-local`, ElasticMQ, Step Functions Local, Cognito Local, AppSync simulator). The template also pinned no image version (`localstack/localstack`), so every scaffolded project silently drifted onto an unmaintained image.

This PR replaces that S3 emulator with [Floci](https://github.com/floci-io/floci) (`floci/floci:1.6.0`): same port (`4566`), same path-style addressing, so application code and `.env` values stay unchanged. Floci starts in tens of milliseconds and idles at ~13 MiB, versus LocalStack’s multi-second startup and hundreds of MB resident. Equivalence is proven with a real-S3 smoke suite that was green against LocalStack first, then re-run unchanged against Floci.

### DIFF
--- a/.cursorrules
+++ b/.cursorrules
@@ -100,7 +100,7 @@ npm run lint                # ESLint check
 npm run format              # Prettier format
 
 # Local Development
-npm run start:localstack    # Start LocalStack services
+npm run offline:docker      # Start local AWS services
 npm run offline             # Start serverless offline
 
 # Database
@@ -110,7 +110,7 @@ npm run ddb:create          # Create DynamoDB tables
 ## Testing Guidelines
 
 - Unit tests: Use Jest with aws-sdk-client-mock for AWS service mocking
-- E2E tests: Use LocalStack for AWS service emulation
+- E2E tests: Use Floci and other local emulators for AWS service emulation
 - Always write tests for new features
 - Maintain test coverage for critical paths
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -24,6 +24,7 @@ module.exports = {
     '.eslintrc.js',
     'prisma',
     '**/*.spec.ts',
+    '**/*.smoke-spec.ts',
     '**/*.test.ts',
     'examples/**',
     // Scaffolding templates are excluded from tsconfig and reference
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -200,7 +200,7 @@ npm run test                # Run unit tests
 npm run test:e2e            # Run E2E tests
 npm run lint                # ESLint check
 npm run format              # Prettier format
-npm run start:localstack    # Start local AWS services
+npm run offline:docker      # Start local AWS services
 npm run offline             # Start serverless offline
 
 # Database
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,42 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 - Harden npm dependencies — resolve high/moderate vulnerabilities via `overrides` (`lodash`, `tmp`, `multer`, `ajv`, `picomatch`, `fast-xml-parser`) and targeted upgrades (`@nestjs/config` ^4.0.4, `@typescript-eslint` v8); root production audit: 0 critical/high ([#445](https://github.com/mbc-net/mbc-cqrs-serverless/pull/445)–[#453](https://github.com/mbc-net/mbc-cqrs-serverless/pull/453))
 - **cli:** Upgrade `multer` to v2.2.0 in scaffolded templates to fix CVE-2026-5079 (DoS via deeply nested multipart field names) ([#463](https://github.com/mbc-net/mbc-cqrs-serverless/pull/463))
 - Add blocking CI gates: `npm audit --omit=dev --audit-level=high` and ESLint startup check run on every PR, failing the build on any production vulnerability ([#448](https://github.com/mbc-net/mbc-cqrs-serverless/pull/448))
+### Breaking Changes
+
+- **infra:** Replace LocalStack with Floci as the local S3 emulator
+
+  LocalStack Community Edition reached end of life in March 2026 — it now requires an auth
+  token and no longer receives security updates. The template pinned no version, so
+  scaffolded projects silently drifted onto an unmaintained image.
+
+  Floci serves S3 on the same port (4566) with the same path-style addressing, so no
+  application code or `.env` value changes. `S3_ENDPOINT`, `LOCAL_S3_PORT`, `S3_REGION`,
+  and `S3_BUCKET_NAME` are unaffected.
+
+  - **Migration:** In `infra-local/docker-compose.yml`, replace the `localstack` service with:
+
+    ```yaml
+      floci:
+        image: floci/floci:1.6.0
+        ports:
+          - '${LOCAL_S3_PORT:-4566}:4566'
+        environment:
+          - FLOCI_DEFAULT_REGION=ap-northeast-1
+          - FLOCI_STORAGE_MODE=persistent
+          - FLOCI_STORAGE_PERSISTENT_PATH=/data
+        volumes:
+          - ./docker-data/floci:/data
+    ```
+
+    Then remove `serverless-localstack` from `package.json` if present.
+
+  - **Data loss warning:** `FLOCI_STORAGE_MODE` defaults to `memory`. Omitting that line
+    discards all buckets on every `docker compose down`, with no error reported. Existing
+    objects under `docker-data/localstack` are not migrated — recreate buckets with
+    `infra-local/scripts/resources.sh`.
+
+  - Existing projects are not required to migrate immediately; the old stack still runs.
+    It runs on an image that no longer receives security patches.
 
 ## [1.3.1](https://github.com/mbc-net/mbc-cqrs-serverless/releases/tag/v1.3.1) (2026-06-02)
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Please be respectful and constructive in all interactions. We welcome contributo
 - Node.js 18 or higher
 - npm 9 or higher
 - Git
-- Docker (for LocalStack)
+- Docker (for Floci and the other local emulators)
 
 ### Development Setup
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -12,7 +12,7 @@ MBC CQRS Serverless is a TypeScript framework for building serverless applicatio
 - **Event Sourcing**: Complete audit trail of all changes
 - **Multi-tenancy**: Built-in tenant isolation
 - **AWS Integration**: Ready-to-use integrations with AWS services
-- **Local Development**: Full local development support with LocalStack
+- **Local Development**: Full local development support with Floci
 
 ### What AWS services does this framework integrate with?
 
@@ -44,7 +44,7 @@ Node.js 18 or higher is required.
 ### How do I set up local development?
 
 1. Copy environment template: `cp .env.example .env`
-2. Start LocalStack: `npm run start:localstack`
+2. Start the local stack: `npm run offline:docker`
 3. Create tables: `npm run ddb:create`
 4. Start server: `npm run offline`
 
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -10,7 +10,7 @@ This guide covers common issues and their solutions when working with MBC CQRS S
 
 **Solutions**:
 1. Ensure Docker is running
-2. Check if ports are available (4566, 4571, etc.)
+2. Check if port 4566 is available
 3. Remove old containers: `docker rm -f floci`
 4. Restart Floci: `npm run offline:docker`
 
@@ -19,9 +19,9 @@ This guide covers common issues and their solutions when working with MBC CQRS S
 **Symptom**: "Table not found" errors when running locally.
 
 **Solutions**:
-1. Ensure Floci is running
+1. Ensure DynamoDB Local is running (`npm run offline:docker`)
 2. Run table creation: `npm run ddb:create`
-3. Verify tables exist: `aws --endpoint-url=http://localhost:4566 dynamodb list-tables`
+3. Verify tables exist: `aws --endpoint-url=http://localhost:8000 dynamodb list-tables`
 
 ### Cognito Authentication Fails Locally
 
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -4,22 +4,22 @@ This guide covers common issues and their solutions when working with MBC CQRS S
 
 ## Local Development
 
-### LocalStack Not Starting
+### Floci Not Starting
 
-**Symptom**: LocalStack container fails to start or services are unavailable.
+**Symptom**: Floci container fails to start or services are unavailable.
 
 **Solutions**:
 1. Ensure Docker is running
 2. Check if ports are available (4566, 4571, etc.)
-3. Remove old containers: `docker rm -f localstack`
-4. Restart LocalStack: `npm run start:localstack`
+3. Remove old containers: `docker rm -f floci`
+4. Restart Floci: `npm run offline:docker`
 
 ### DynamoDB Tables Not Created
 
 **Symptom**: "Table not found" errors when running locally.
 
 **Solutions**:
-1. Ensure LocalStack is running
+1. Ensure Floci is running
 2. Run table creation: `npm run ddb:create`
 3. Verify tables exist: `aws --endpoint-url=http://localhost:4566 dynamodb list-tables`
 
@@ -130,7 +130,7 @@ beforeEach(() => {
 **Symptom**: E2E tests fail with connection errors.
 
 **Solutions**:
-1. Start LocalStack before running tests
+1. Start the local stack before running tests: `npm run offline:docker`
 2. Ensure test database is clean
 3. Check test configuration in `jest.config.js`
 
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -215,7 +215,7 @@ npm run lint
 npm run format
 
 # Local Development
-npm run start:localstack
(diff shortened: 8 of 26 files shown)


koichimurakami (commented):
## Review: Changes requested (please address the 2 HIGH items before merging)

As a response to the LocalStack Community Edition EOL, I think the approach here is sound. These points in particular are real improvements:

- Only S3 — the single service LocalStack was actually used for — is replaced; the rest of the emulator setup is left untouched.
- The image is pinned to `floci/floci:1.6.0`. The old `localstack/localstack` had no tag pinned, which is exactly the drift the PR description calls out.
- A smoke test was added as evidence of equivalence.
- `npm run start:localstack`, referenced throughout the docs, never actually existed as a script in `package.json`. Replacing it with `offline:docker` also fixes that pre-existing documentation bug.

For this review I ran Floci 1.6.0 and LocalStack 3.8 side by side and verified the behavior on real containers. The claims in the PR description hold up against the measurements:

| Metric | Floci 1.6.0 | LocalStack 3.8 |
|---|---|---|
| Startup time | 0.222 s | ~3 s |
| Memory (idle) | 15.2 MiB | 425.6 MiB |
| Smoke test | 4/4 PASS | 4/4 PASS |

I confirmed the smoke test passes against both. The test itself is sound, and the equivalence claim is backed by measurement for the four paths it covers.

That said, there are two items that need to be addressed before merging.

---

## HIGH-1: The new smoke test is never executed by CI

Nothing invokes the `test:s3-smoke` script added to `packages/core/package.json`.

- The `e2e_tests` job in `.github/workflows/ci-cd.yml` only runs `npx lerna run test:e2e --scope=@mbc-cqrs-serverless/core`.
- `test:e2e` uses `test/jest-e2e.json`, whose `testRegex` is `.e2e-spec.ts$`, so it does not match `s3.smoke-spec.ts`.
- Searching the entire workflow directory returns no occurrence of `smoke`, and the root `package.json` has no invocation either.
- `jest-s3-smoke.json` has no `globalSetup` / `globalTeardown` (unlike `jest-e2e.json`), so it does not start Floci on its own — it assumes `offline:docker` was already run manually.

On top of that, the existing e2e specs (`health` / `publish-*`) do not touch S3 at all. As a result, **S3 is never actually exercised in CI, either before or after this PR.**

There is nothing wrong with the test content — I confirmed 4/4 PASS on real containers — the CI wiring is simply missing. As it stands, the verification behind "the suite was green against LocalStack first, then re-run unchanged against Floci" becomes a one-time manual step that nobody reproduces after merge. Future Floci image bumps or docker-compose changes would leave CI green regardless.

**Suggested fixes (either one)**

1. Rename `s3.smoke-spec.ts` to `s3.e2e-spec.ts` and fold it into the existing E2E suite. Since this PR enables floci in `packages/core/test/infra-local/docker-compose.yml`, the stack started by the existing `globalSetup` already includes Floci. This is the smallest change.
2. If you prefer to keep it separate, add `docker compose up -d --wait floci` → `npm run test:s3-smoke` → a compose down guarded by `if: always()` to CI.

---

## HIGH-2: LocalStack's `EXTRA_CORS_ALLOWED_ORIGINS=*` was removed with no replacement

`packages/directory/src/directory-file.service.ts` issues presigned GET / PUT URLs via `getSignedUrl` (`genViewUrl`, and `genUploadDirectoryUrl` with `ACL: 'private'`). This is the browser-to-S3 direct access path. Production CDK sets CORS explicitly on the relevant buckets, but locally that role was filled by LocalStack's `EXTRA_CORS_ALLOWED_ORIGINS=*`. The new floci service definition has no equivalent.

I verified this on real containers, and **it is a regression.** OPTIONS preflight against a bucket with no CORS configuration:

| Environment | PUT preflight | GET preflight |
|---|---|---|
| LocalStack 3.8 (previous template config) | 200 / `Access-Control-Allow-Origin: http://localhost:3000` | 200 / same |
| Floci 1.6.0 (as configured in this PR) | **403** `CORSResponse: This CORS request is not allowed.` | **403** same |

Reproducing the exact `DirectoryFileService` flow (issue presigned PUT → real PUT → issue presigned GET → real GET):

- The real PUT and GET themselves succeed with 200 on Floci. Both `ACL: 'private'` and `ResponseContentDisposition` are accepted correctly.
- However, preflight fails with 403, and the actual responses carry no `Access-Control-Allow-Origin` header either.
- Therefore **the browser path does not work** — the actual request is never sent once preflight fails.
- Server-side (Node.js) usage is unaffected.

The new smoke test only exercises the Node.js `S3Client` path, using neither an `Origin` header nor a presigned URL, so it cannot detect this regression.

**Suggested fixes (both verified on real containers to yield 200 + ACAO across all paths)**

1. **Add `PutBucketCors` to `infra-local/scripts/resources.sh` (recommended).**
   Floci supports `put-bucket-cors` / `get-bucket-cors` correctly. After configuring it, I confirmed that both preflight and the actual PUT / GET requests return 200 with `Access-Control-Allow-Origin`. It is a documented API, and it has the added benefit of keeping local behavior aligned with the production CDK CORS configuration.
2. Add `FLOCI_SECURITY_EXTRA_CORS_ALLOWED_ORIGINS=*` to docker-compose.
   This restores parity with the old setup in a single line (verified: 204 with `Access-Control-Allow-Origin: *`). However, **this variable does not appear in Floci's README configuration table.** It could change without notice, which is the same class of risk as the unpinned-tag situation that left us stranded on LocalStack.

Note that **passing `EXTRA_CORS_ALLOWED_ORIGINS` (the LocalStack name) as-is does not work** — it still returned 403. It is not covered by Floci's automatic LocalStack environment variable translation.

---

## MEDIUM

### 1. Placement of the CHANGELOG entry

The new `### Breaking Changes` block is inserted at the end of the already-released `## [1.3.2]` (2026-06-27) section, right after the Security items. 1.3.2 was followed by 1.3.4 / 1.3.5 / 1.4.0, all already released, while this change is not yet merged. As written, the published CHANGELOG will read as though the LocalStack-to-Floci migration shipped in v1.3.2.

The subsection order within that section also becomes Features / Bug Fixes / Security / Breaking Changes, which does not match the format used in 1.2.4, where Breaking Changes comes first.

**Suggested fix**: Add a new `## [Unreleased]` section at the top and move the entry there.

### 2. The emulator's exposed surface widens from S3 to every service

Floci has no equivalent of LocalStack's `SERVICES=s3` service selection. I verified that the health endpoint reports **all 72 services as `running`**, and the startup log lists 71 services under `Enabled services`. With dummy credentials, `dynamodb list-tables`, `iam list-users`, and `secretsmanager list-secrets` all respond normally on 4566.

The `0.0.0.0` port binding is unchanged from the LocalStack setup, so the binding itself is not a regression — but the range of exposed APIs widens from S3 alone to every service.

**Suggested fix**: If host-only access is sufficient, restrict it to loopback with `127.0.0.1:${LOCAL_S3_PORT:-4566}:4566`.

### 3. The significance of `FLOCI_STORAGE_MODE` is not documented in the compose file itself

As this PR's own CHANGELOG states, `FLOCI_STORAGE_MODE` defaults to `memory`, and omitting it discards all data on every restart. I confirmed this on real containers:

- With this PR's configuration (`persistent`), buckets and objects survive `docker compose down` → `up`.
- With `FLOCI_STORAGE_MODE` omitted, everything was gone after a restart, returning `NoSuchBucket`. The startup log does print an INFO line, `Storage: memory  Path: /app/data` — but it is INFO rather than WARN, and no error is raised when the data is discarded.

The current docker-compose has no comment indicating how load-bearing that single line is. This is a template that gets copied into every new project via `mbc new`, so if the line is ever dropped or rewritten into the `${...:-default}` form used by the other port settings, Floci will keep starting without complaint.

**Suggested fix**: Add a comment directly above that line explaining the consequence of omitting it.

### 4. The smoke test's credentials and bucket name differ from the real application

`packages/core/test/e2e/s3.smoke-spec.ts` defaults to `DUMMYIDEXAMPLE` / `DUMMYEXAMPLEKEY` for credentials and uses `e2e-s3-smoke` as the bucket name. The actual application `.env`, meanwhile, uses `local` / `local` and `S3_BUCKET_NAME=local-bucket`. The path exercised in production is therefore not the one under test.

**Suggested fix**: Either use the same defaults as `.env` (or `.env.example`), or initialize through the same configuration source as `S3Service`.

---

## LOW

1. **`packages/cli/templates/.env.local` (line 27)**: The comment `# LocalStack (S3)` is still there. Since this file is copied by `mbc new`, every newly generated project will carry a reference to a LocalStack that is no longer present. Please update it to `# Floci (S3)`.
2. **`packages/mcp-server/skills/mbc-debug/SKILL.md`**: "Floci is S3-only on :4566" is accurate as a description of how this project uses it, but inaccurate as a description of Floci itself, which supports the full range of AWS services. Consider rewording to avoid the misunderstanding.
3. **`packages/core/src/config/service-config.spec.ts` and `packages/core/src/integration/aws-sesv2-behavior.spec.ts`**: References to LocalStack remain. These are only test names and variable names with no behavioral impact, so addressing them is optional.

---

## How this was verified

I ran Floci 1.6.0 and LocalStack 3.8 concurrently in Docker and checked each finding against real containers.

- Floci was started with the exact configuration from this PR's `docker-compose.yml`.
- LocalStack was started with the pre-migration template configuration (`SERVICES=s3`, `EXTRA_CORS_ALLOWED_ORIGINS=*`, and so on).
- The smoke test was extracted from the PR diff and run against both endpoints.

Type checking, unit tests, and the build were not run: the repository has concurrent work in progress, so I could not change git state to check out the PR branch. That said, by cross-checking the `testRegex` in `jest-e2e.json` against the `exclude` in `test/tsconfig.build.json`, I did statically confirm that **this PR will not break the existing CI.**

koichimurakami (commented):
## Review: changes requested (1 HIGH, 2 MEDIUM, 5 LOW)

Reviewed at `51e064a`. Thanks for the thorough migration and the smoke suite. CI is green, and I confirmed that `s3.e2e-spec.ts` runs in the E2E job. I re-checked every finding below by reproducing it (AWS CLI, Floci 1.6.0 in Docker, npm 10/11).

### HIGH

**1. `resources.ps1`: the CORS rule is never applied on Windows** (`packages/cli/templates/infra-local/scripts/resources.ps1`)

`resources:win32` runs the script with `powershell -File`, which is Windows PowerShell 5.1. In 5.1, `Set-Content -Encoding utf8` writes a UTF-8 **BOM**. The AWS CLI does not strip it: `paramfile.get_file` opens the file with the locale encoding, not `utf-8-sig`. So the call fails with:

```
Error parsing parameter '--cors-configuration': Expected: '=', received: '﻿'
```

(Reproduced by passing a BOM-prefixed copy of the same JSON: with the BOM it fails as above, without it the JSON parses.)

`$LASTEXITCODE` is not checked, so the script carries on and the bucket ends up with no CORS rule. Floci answers the browser preflight with **403** when no rule is set (200 once one is set, verified on 1.6.0). As a result, presigned upload and view URLs stop working in the browser on Windows.

Suggested fix:

```powershell
[System.IO.File]::WriteAllText($corsFile, $corsConfiguration, (New-Object System.Text.UTF8Encoding $false))
...
aws --endpoint-url=http://localhost:$s3Port s3api put-bucket-cors --bucket $env:S3_BUCKET_NAME --cors-configuration "file://$corsFile"
if ($LASTEXITCODE -ne 0) { throw "put-bucket-cors failed ($LASTEXITCODE)" }
```

### MEDIUM

**2. `package-lock.json` is out of sync for npm 11**

With npm 11.16 (the npm bundled with Node 24), `npm ci` fails on this branch:

```
npm error Invalid: lock file's picomatch@4.0.5 does not satisfy picomatch@4.0.7
```

`develop` passes `npm ci` with both npm 10 and npm 11. This branch passes only with npm 10. The lock also moves `node_modules/filelist/node_modules/brace-expansion` from 5.0.8 **down** to 2.1.7, even though the override is `"brace-expansion@2": "5.0.12"`. CI doesn't catch this because the `Sync lockfile` step (`npm install --package-lock-only`) regenerates the lock before `npm ci`. Contributors on Node 24 cannot do a clean install.

Suggested fix: regenerate the lock (delete the stale nested entries, run `npm install --package-lock-only`) and confirm that `npm ci` passes under both npm 10 and npm 11.

**3. The CHANGELOG migration guide leaves out the CORS step** (`CHANGELOG.md`, Breaking Changes)

The migration steps only swap the compose service and remove `serverless-localstack`. LocalStack used to allow CORS through `EXTRA_CORS_ALLOWED_ORIGINS=*`, and Floci does not (403 on preflight, see #1). An existing project that follows the guide as written loses browser uploads and views through presigned URLs. Please add a step that applies the bucket CORS rule, either by copying the new block into `resources.sh`/`resources.ps1` or by running `put-bucket-cors` manually.

### LOW

4. **`.eslintrc.js`**: the new `'**/*.smoke-spec.ts'` ignore pattern matches no file. The spec was renamed to `s3.e2e-spec.ts` later in this PR. The pattern can be dropped.
5. **CHANGELOG vs. the compose file you ship**: the migration example uses `FLOCI_STORAGE_PERSISTENT_PATH=/data` with the bind mount `./docker-data/floci:/data`. The template and the test stack now ship the named volume `floci-data:/app/data`. Both configurations work (I verified that the bind-mount variant persists across `down`/`up`), but they should match to avoid confusion.
6. **`packages/cli/templates/.env.local:29`**: the comment still says `# LocalStack (S3)`.
7. **Test-stack state carries over between local runs**: `test/infra-local` runs Floci in `persistent` mode on a named volume, and `stop.sh` only runs `docker compose down`. If the smoke suite is interrupted before `afterAll`, the bucket `e2e-s3-smoke` is left behind. On the next local run, `beforeAll` then fails with `BucketAlreadyOwnedByYou` (verified on Floci 1.6.0). CI is unaffected. Consider `FLOCI_STORAGE_MODE=memory` for the test stack, or tolerate `BucketAlreadyOwnedByYou` in `beforeAll`.
8. **Overrides**: the new exact pins (`js-yaml 4.3.2`, `fast-uri 3.1.8`, `brace-expansion@2 5.0.12`) will again become "one release before the fix" the next time an advisory lands, so caret lower bounds (e.g. `^4.3.2`) would age better. The same overrides are also being changed in #504. That change is outside this PR's scope and the two will conflict, so it may be better to keep the dependency bumps in one PR.

### Verification

| Check | Result |
|---|---|
| CI (Unit / E2E 18–24.x / Extended Matrix / Trivy) | pass |
| `npm ci` (npm 10 / npm 11.16) | pass / **fail** (develop: pass / pass) |
| `npm audit --omit=dev --audit-level=high` | pass (overall high count 67 → 54) |
| prettier on the changed TS files, mcp-server `skills.spec.ts` (64/64), `tsc -p test/tsconfig.build.json` | pass |
| Floci 1.6.0 manual check | preflight 403 without CORS / 200 with CORS; bind-mount persistence OK |

koichimurakami (approved):
## Re-review at `dbe40e6`: approved with follow-ups (0 HIGH, 2 MEDIUM, 3 LOW)

Thanks for the quick turnaround. The HIGH from the previous review is fixed: `resources.ps1` now writes the CORS file without a BOM and checks `$LASTEXITCODE`. I could only review that change by reading it, since I have no PowerShell here to run it. The ESLint ignore pattern, the mismatch between the CHANGELOG example and the shipped compose file, and the `.env.local` comment are also fixed. CI is green (14 pass).

I'm approving. The items below are not blockers, but please handle the two MEDIUMs in a follow-up PR (or here, if you prefer).

### Status of the previous findings

| Previous finding | Status |
|---|---|
| HIGH `resources.ps1` BOM / unchecked exit code | ✅ fixed |
| MEDIUM lockfile out of sync for npm 11 | ⏳ still open |
| MEDIUM CHANGELOG migration guide has no CORS step | ⏳ still open |
| LOW `.eslintrc.js` dead pattern | ✅ fixed |
| LOW CHANGELOG example vs. shipped compose | ✅ fixed |
| LOW `.env.local` comment | ✅ fixed |
| LOW test-stack `BucketAlreadyOwnedByYou` | ⏳ still open |
| LOW exact pins / overlap with #504 | ⏳ still open |

### MEDIUM (follow-up recommended)

1. **`package-lock.json` still fails `npm ci` on npm 11.** Re-checked at `dbe40e6` with npm 11.16: `npm error Invalid: lock file's picomatch@4.0.5 does not satisfy picomatch@4.0.7`. `develop` passes under the same npm. CI hides this because the `Sync lockfile` step regenerates the lock first. Please regenerate the lock so that `npm ci` passes under both npm 10 and npm 11.
2. **The CHANGELOG migration guide still has no CORS step.** An existing project that follows the guide keeps its old `resources.sh`/`resources.ps1`, so its bucket gets no CORS rule. Floci then answers the browser preflight with 403, which breaks presigned upload and view URLs. Please add a step that applies the new CORS block, or a one-off `put-bucket-cors` command.

### LOW

3. **`resources.sh` (new)**: `put-bucket-cors` is not exit-checked, so the script now behaves differently from `resources.ps1`. With a stubbed `aws` that fails only the CORS call, the script still exits 0. The impact is small: the AWS CLI error is still printed, and in `offline:sls:default` the `(resources.sh && trigger_ddb_stream.sh) &` group runs in the background, so `serverless offline` starts either way. Still, adding `|| { echo "put-bucket-cors failed" >&2; exit 1; }` would make the two scripts consistent.
4. **Test stack (still open)**: if a local run is interrupted before `afterAll`, the persistent `floci-data` volume keeps `e2e-s3-smoke`, and the next local run fails in `beforeAll` with `BucketAlreadyOwnedByYou`. CI is unaffected.
5. **Overrides (still open)**: `js-yaml` / `fast-uri` / `brace-expansion@2` are still exact pins, and #504 (still open) changes the same lines. Whichever PR merges second will need a rebase.

PR #504, opened 2026-08-20
fix(deps): bump security overrides to fixed releases to unblock the audit gate

koichimurakami:
## Summary

The blocking CI gate `npm audit --omit=dev --audit-level=high` (ci-cd.yml:70-71) is failing on `develop` with **7 high advisories**, which is why every PR's Unit Test job is red and `e2e_tests` is skipped.

No commit caused this. The `Sync lockfile` step (`npm install --package-lock-only`) re-resolves the tree on every run, so newly published advisories turn the build red on their own — `develop`'s last green CI was 2026-08-02.

The root cause is that three existing overrides had each drifted to **exactly one patch below** the advisory's fixed release, so they were actively pinning vulnerable versions in place:

| override | before | after | vulnerable range |
|---|---|---|---|
| `js-yaml` | `4.3.0` | `^4.3.1` | `4.0.0 - 4.3.0` |
| `fast-uri` | `3.1.4` | `^3.1.5` | `3.0.0 - 3.1.4` |
| `brace-expansion@2` | `5.0.8` | `5.0.9` | `4.0.0 - 5.0.8` |
| `ip-address` | *(none)* | `^10.3.1` | `<= 10.3.0` |

Only these four are root causes. `@nestjs/swagger`, `ajv` and `minimatch` were flagged solely as downstream effects (`via`) and clear once the roots are fixed. All four bumps stay within the same major.

### Why `brace-expansion@2` keeps its selector

A bare `brace-expansion` key also clears the audit, and it looks tidier — but it is wrong here, and the CI would not tell you.

brace-expansion 5.x no longer exports a callable from its CJS entry (`require()` yields `{ EXPANSION_MAX, EXPANSION_MAX_LENGTH, expand }`). ESLint's nested minimatch 3.1.5 does `var expand = require('brace-expansion')` and then `expand(pattern)`, so a bare override makes **every braced pattern** throw:

```
TypeError: expand is not a function
    at Minimatch.braceExpand (node_modules/eslint/node_modules/minimatch/minimatch.js:271:10)
```

The `ESLint startup check` step (added in #446 to catch exactly this class of breakage) does **not** brace-expand, so it passes — the failure would ship silently. Keeping the `@2` selector from #482 leaves ESLint on its own 1.1.16 copy while still moving the hoisted production copy to 5.0.9.

### Why the lockfile entries are dropped

The stale `node_modules/{brace-expansion,js-yaml,fast-uri,ip-address}` entries are removed so the incremental re-resolve re-evaluates them. Without that, `npm install --package-lock-only` keeps the previously pinned versions and the gate stays red even with the new overrides. Lockfile churn is 150 lines.

Caret ranges replace the two exact pins so future patch releases are picked up by the same re-resolution step that surfaced this, instead of freezing the tree one patch behind the next advisory.

## Verification

Against `origin/develop`, with a clean `npm ci --ignore-scripts`:

| check | before | after |
|---|---|---|
| `npm audit --omit=dev --audit-level=high` (blocking) | **exit 1** — 7 high | **exit 0** — 0 high |
| `ESLint startup check` (blocking) | pass | **pass** |
| ESLint's minimatch 3.1.5 brace expansion (runtime; not covered by CI) | works | **works** — resolves brace-expansion 1.1.16 |
| hoisted production `brace-expansion` | 5.0.8 (vulnerable) | **5.0.9** |
| `npm run lint` | — | **runs normally** — 26 pre-existing rule violations, 0 crashes |
| `npm run test:cov` | — | **167/168 suites, 3658 tests pass** |

The one failing suite (`packages/import/src/event/csv-import.sfn.event.handler.spec.ts`) fails with `TS1261: ... differs from file name ... only in casing` on `@types/JSONStream` vs `@types/jsonstream` — a macOS case-insensitive-filesystem artifact, unrelated to this change, and absent on the Linux CI runners.

## Note

This does not touch the moderate/low advisories (`hono`, `file-type`, `@nestjs/*`, `body-parser`), which the gate tolerates at `--audit-level=high`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7697,6 +7697,17 @@
         "node": ">= 6.0.0"
       }
     },
+    "node_modules/@serverless/platform-client/node_modules/brace-expansion": {
+      "version": "1.1.18",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.18.tgz",
+      "integrity": "sha512-Edep/X9fGqVNmzKBVsDYIOtD+z1tuezV70LBjdCst9Tqu76lsnvRiZ6oTic1n+/BIwX6QDGAO94PN4N2SADvtw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "balanced-match": "^1.0.0",
+        "concat-map": "0.0.1"
+      }
+    },
     "node_modules/@serverless/platform-client/node_modules/https-proxy-agent": {
       "version": "5.0.1",
       "resolved": "https://registry.npmjs.org/https-proxy-agent/-/https-proxy-agent-5.0.1.tgz",
@@ -10637,6 +10648,17 @@
         "node": ">= 6"
       }
     },
+    "node_modules/archiver-utils/node_modules/brace-expansion": {
+      "version": "1.1.18",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.18.tgz",
+      "integrity": "sha512-Edep/X9fGqVNmzKBVsDYIOtD+z1tuezV70LBjdCst9Tqu76lsnvRiZ6oTic1n+/BIwX6QDGAO94PN4N2SADvtw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "balanced-match": "^1.0.0",
+        "concat-map": "0.0.1"
+      }
+    },
     "node_modules/archiver-utils/node_modules/glob": {
       "version": "7.2.3",
       "resolved": "https://registry.npmjs.org/glob/-/glob-7.2.3.tgz",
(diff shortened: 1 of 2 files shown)


koichimurakami:
Closing this PR: #503 has superseded it.

#503 (merged into `develop` on 2026-09-23) bumped the same overrides to newer fixed releases than this PR does:

| override | this PR | `develop` now |
|---|---|---|
| `js-yaml` | `^4.3.1` | `4.3.2` |
| `fast-uri` | `^3.1.5` | `3.1.8` |
| `brace-expansion@2` | `5.0.9` | `5.0.12` |
| `ip-address` | `^10.3.1` | `^10.7.2` |

- This branch now conflicts with `develop` in `package.json` and `package-lock.json`. Resolving the conflict in favour of this branch would move the overrides back to older versions.
- The goal of this PR is already met on current `develop` (`7981a98`): `npm audit --omit=dev --audit-level=high` exits 0 with 0 high findings. That holds both with the committed lockfile and after the CI-style `npm install --package-lock-only` re-resolve, and the latest `develop` CI run is green.

The one idea not carried over is switching the exact pins to caret ranges, so that future patch releases are picked up automatically. If we want that, it should be a small new PR against current `develop`. It should keep the `brace-expansion@2` selector, for the ESLint/minimatch reason explained above.

PR #505, opened 2026-09-16
fix(directory): context option

tudang-z:


koichimurakami (approved):
## Review: approved (0 HIGH/MEDIUM, 1 LOW)

Reviewed at `640ddb8`. Thanks! Widening `opts` from `{ invokeContext: IInvoke }` to `ICommandOptions` lets callers pass `source` / `requestId` through to the commands. The service already forwards `opts` unchanged to `publishAsync` / `publishPartialUpdateAsync`, so no further wiring is needed.

It is backward compatible: `ICommandOptions` still requires `invokeContext`, so existing callers that pass `{ invokeContext }` (the controller, for example) compile unchanged, and subclasses that override these methods with the old parameter type still type-check. No `{ invokeContext: IInvoke }` signature is left in the directory package.

### LOW (optional follow-up)

- **Tests / CHANGELOG:** there is no test that a `source` / `requestId` given to `create` / `update` / `remove` etc. actually reaches `commandService.publishAsync` / `publishPartialUpdateAsync`. A small case in `directory.service.spec.ts` would lock this in. A CHANGELOG line along the lines of "directory: service methods accept `ICommandOptions` (backward compatible)" would also help consumers find the new capability.

### Verification

| Check | Result |
|---|---|
| CI | pass |
| `tsc --noEmit` (packages/directory, with dependencies built) | pass |
| `jest packages/directory` | 2 suites / 59 tests pass |
| prettier / eslint on `directory.service.ts` | pass |
| Merge with current `develop` (after #506–#508) | clean, no conflicts |

